PR #7, opened 2023-02-04
Change all "InPlace" functions to return unit

cknitt:


glennsl:
Is it guaranteed that expressions using these functions will always return unit/undefined? I've done some testing, with GADTs and such, to try to trip the compiler up, but it seems to always hold. I've not seen this behaviour documented anywhere though.

cristianoc:
Why does it matter what they return? There's no way to inspect values of type unit.

glennsl:
In rescript no, but in JS it's done all the time. Functions that take a callback and behave differently depending on what the callback returns, for example.

PR #16, opened 2023-02-09
Add JSON decode functions from Js.Json

cknitt:
... and move classify into separate submodule + add .resi.

Merge #15 first.

Note: I made it Json.Decode.float (Js.Json calls it decodeNumber) and am not providing a Json.Decode.int on purpose.

### DIFF
--- a/src/RescriptCore.mjs
+++ b/src/RescriptCore.mjs
@@ -15,8 +15,6 @@ var $$Error;
 
 var Float;
 
-var Global;
-
 var Int;
 
 var $$BigInt;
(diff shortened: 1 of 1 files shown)


src/Core__JSON.resi
```diff
@@ -0,0 +1,55 @@
+type t = Js.Json.t
+
+type jsonReviver
+external asJsonReviver: 'a => jsonReviver = "%identity"
+type jsonReplacer
+external asJsonReplacer: 'a => jsonReplacer = "%identity"
```
glennsl, line 6: What's the point of these? They look unsafe, but are not marked as such. And if they can really be anything, why not just allow anything to be passed directly?

Also, as I understand, `jsonReviver` can really only be a function. So why isn't it a function?

PR #24, opened 2023-02-10
Add (immutable) sort to Array

cknitt:
* sort is the immutable version of the JS Array.sort function
* ~~stableSort, stableSortInPlace are the implementation from Belt~~
* Use `Array.sortInPlace` for `List.sort`

### DIFF
--- a/src/Core__Array.mjs
+++ b/src/Core__Array.mjs
@@ -21,6 +21,12 @@ function lastIndexOfOpt(arr, item) {
   
 }
 
+function sort(arr, cmp) {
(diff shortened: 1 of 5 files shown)


cknitt:
It seems we do not need our own stable sort implementation.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

> Since version 10 (or ECMAScript 2019), the [specification](https://tc39.es/ecma262/#sec-array.prototype.sort) dictates that Array.prototype.sort is stable.

PR #27, opened 2023-02-12
Documentation: String

aspeddro:


src/Core__String.resi
```diff
@@ -0,0 +1,735 @@
...
+```
+*/
+@send
+external endsWith: (string, string) => bool = "endsWith"
+
+/**
+`endsWithFrom(str, ending, len)` returns `true` if the first len characters of
+`str` end with `ending`, `false` otherwise. If `len` is greater than or equal
+to the length of `str`, then it works like `endsWith`. (Honestly, this should
+have been named endsWithAt, but oh well).
```
aspeddro, line 185: Should be renamed to `endsWithAt`?

zth: No let's no change it now. But remove that part of the comment perhaps? Doesn't feel like a comment like that belongs in the docstring.

src/Core__String.resi
```diff
@@ -0,0 +1,795 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA. */
+
+/*** Provide bindings to JS string.*/
```
zth, line 25:
```suggestion
/*** Functions for interacting with JavaScript strings.
See: [`String`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String).*/
```

zth (commented):
Looking good! Just a few small comments/suggestions. Also, could we ensure that nothing is prefixed with `Belt` in here, so we refer to the right modules? I saw that some example code had references to `Belt.Int` for example, when it should be just `Int`.

src/Core__String.resi
```diff
@@ -0,0 +1,1017 @@
...
+String.padEnd("abc", 1, "") == "abc"
+```
+*/
+@send
+external padEnd: (string, int, string) => string = "padEnd"
+
+// TODO: add docs
+@get_index external getSymbol: (string, Core__Symbol.t) => option<'a> = ""
+@get_index external getSymbolUnsafe: (string, Core__Symbol.t) => 'a = ""
+@set_index external setSymbol: (string, Core__Symbol.t, 'a) => unit = ""
```
aspeddro, line 999: It remains to document these three.

src/Core__String.resi
```diff
@@ -0,0 +1,978 @@
...
+See [`String.split`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) on MDN.
+
+## Examples
+
+```rescript
+String.splitByRegExpAtMost("Hello World. How are you doing?", %re("/ /"), ~limit=3) == [
+      Some("Hello"),
+      Some("World."),
+      Some("How"),
+    ]
```
zth, line 751: Indentation.

src/Core__String.resi
zth: 👍

zth (commented):
Great stuff! 2 small comments, and then docstrings for the 3 missing things that you've already noted, and then I believe we're good to go.

zth:
@aspeddro is this ready to go?

src/Core__String.resi
aspeddro: I've never used `Symbol` in JS. Is there any example case for these bindings?

```res
String.getSymbol("name", Symbol.make("name"))
```

Compile to:
```js
"name"[Symbol("name")]
```

zth: Yeah I don't know, maybe symbols in strings isn't a thing and this binding is unnecessary?

I say we can skip docstrings for this now, and add a separate issue to investigate if these are to be removed.

aspeddro:
> @aspeddro is this ready to go?

yes

src/Core__String.resi
aspeddro: See https://github.com/rescript-association/rescript-core/issues/80

zth (approved):

PR #30, opened 2023-02-13
Add locale and formatting options to localeDateString, localeString and localTimeString functions

dkirchhof:
I didn't implement the full spec, but added hopefully the most important options.   
I took this article as source https://medium.com/swlh/use-tolocaledatestring-to-format-javascript-dates-2959108ea020


zth (approved):
Great stuff, thank you very much!

2 small things I wonder about:
1. Could you add a comment somewhere on what things from the spec you left out? In case someone wants to pick that up.
2. Could you add docstrings for the things you've added? I know the rest of the file doesn't have docstrings yet (we'll fix that), but it'd be nice to include it directly for anything we add.

dkirchhof:
@zth I tried to generate an interface file to add the doc strings there, but it seems that the command `bsc -i` doesn't work anymore?

For my additions I could write it by hand but I don't want to "hide" all other functions.
How do you generate your resi files?

zth:
> @zth I tried to generate an interface file to add the doc strings there, but it seems that the command `bsc -i` doesn't work anymore?
> 
> For my additions I could write it by hand but I don't want to "hide" all other functions. How do you generate your resi files?

There's a command in the VSCode extension that can generate interface files, I use that.

dkirchhof:
Okay, thanks :) I will install vscode for that :roll_eyes: 

Todo for me: I have to figure out, how I can implement it in the vim plugin.

zth:
Yeah that would be good to have in vim too for sure.

dkirchhof:
So, I added the doc strings. Wasn't sure about formatting. Existing documentation is indented, the documentation in open PRs isn't.

dkirchhof:
> Yeah that would be good to have in vim too for sure.

FYI https://github.com/rescript-lang/vim-rescript/issues/66

src/Core__Date.resi
```diff
@@ -0,0 +1,294 @@
...
+@send external toTimeString: t => string = "toTimeString"
+
+/** 
+  `toLocaleDateString(date)`
+
+  Converts a JavaScript date to a localized date string. It will use the current locale.
+
+  ## Examples
+  ```rescript
+  Date.make()->Date.toLocaleDateString->Console.log
```
zth, line 176: Could you also add a comment with what the output will be when you log to the console here? Or rather, what the output is when you do it now as you're adding the docstrings.

This goes for all the examples.

src/Core__Date.resi
```diff
@@ -0,0 +1,294 @@
...
+    - dayPeriod
+    - calendar
+    - numberingSystem
+    - localeMatcher
+    - timeZone
+    - hour12
+    - hourCycle
+    - formatMatcher
+
+  See full spec at https://tc39.es/ecma402/#datetimeformat-objects
```
zth, line 22: You can remove the indentation here.

zth (commented):
Great docstrings! Just a few small adjustments and this is good to go. Thank you!

zth (approved):
Fantastic, great work, thank you! Merging.

Would you potentially be up for adding similar docstrings for the rest of Date in a separate PR?

dkirchhof:
> Fantastic, great work, thank you! Merging.
> 
> Would you potentially be up for adding similar docstrings for the rest of Date in a separate PR?

I was just thinking about it too. I will ;)

PR #32, opened 2023-02-14
Docs for Type module

zth:
Basic docstrings for the `Type` module.

### DIFF
--- a/src/Core__Type.resi
+++ b/src/Core__Type.resi
@@ -1,12 +1,62 @@
+/***
+Utilities for classifying the type of JavaScript values at runtime.
+*/
+
+/**
+The possible types of JavaScript values.
+*/
 type t = [#undefined | #object | #boolean | #number | #bigint | #string | #symbol | #function]
 
+/**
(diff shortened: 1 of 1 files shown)


src/Core__Type.resi
```diff
@@ -1,12 +1,62 @@
...
+
+  /**
+  An abstract type representing a JavaScript function.
+
+  See [`function`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) on MDN.
+  */
   type function
+
+  /**
+  An abstract type representing an unknown JavaScript object.
```
cknitt, line 44: Why "unknown"?

src/Core__Type.resi
```diff
@@ -17,5 +67,17 @@ module Classify: {
     | Function(function)
     | Symbol(symbol)
 
+  /**
+`classify(anyValue)`
+Classifies any value to its underlying JavaScript value.
```
cknitt, line 72: The above variant is not really the underlying JS value.

PR #36, opened 2023-02-16
Docs for List

aspeddro:
- Add examples header `## Examples`
- Format code blocks
- Remove indentation
- I rewrote some comments

### DIFF
--- a/src/Core__List.resi
+++ b/src/Core__List.resi
@@ -23,129 +23,158 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA. */
 
 /**
-  Collection functions for manipulating the `list` data structures, a singly-linked list.
+Collection functions for manipulating the `list` data structures, a singly-linked list.
 
-  **Prefer Array** if you need any of the following:
+**Prefer Array** if you need any of the following:
 
-  - Random access of element
-  - Better interop with JavaScript
-  - Better memory usage & performance.
+- Random access of element
+- Better interop with JavaScript
+- Better memory usage & performance.
 */
 /** `'a t` is compatible with built-in `list` type */
 type t<'a> = list<'a>
 
 /**
-  Returns the length of a list.
+`length(list)` returns the length of `list`.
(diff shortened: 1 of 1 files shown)


src/Core__List.resi
```diff
@@ -23,129 +23,158 @@
...
+
+```rescript
+List.length(list{1, 2, 3}) // 3
+```
 */
 let length: t<'a> => int
 
-/** **See** [`length`](##length) */
+/**
+`size(list)`. **See** [`length`](##length)
```
aspeddro, line 49: There are several comments using this format:

```
[`length`](##length)
```

In VSCode this is a link that opens a file that doesn't exist.

image


How should we reference functions?

aspeddro: A proposal:

```
[`function`](ModuleName.function)
```

For function in same module:

```
[`function`](function)
```

zth: Right, that's an internal markdown link. We haven't 100% landed on the format yet (it's dependent on the documentation extraction, which we're working on in parallel). Let's revisit this a bit later, it's easy to fix once we have the format we want down.

src/Core__List.resi
```diff
@@ -23,129 +23,158 @@
...
+List.head(list{}) // None
+List.head(list{1, 2, 3}) // Some(1)
+```
 */
 let head: t<'a> => option<'a>
 
 /**
-  Same as [head](#head), but raises an exception if `someList` is empty. Use
-  with care.
+`headExn(list)` same as [`head`](#head), but raises an exception if `list` is
```
zth, line 73: Should we add the same `## Exceptions` section here that you added in another PR?

zth (approved):
Just a comment!

src/Core__List.resi
aspeddro: Done

src/Core__List.resi
aspeddro: Ok

zth (approved):

PR #37, opened 2023-02-16
Docs for Int

aspeddro:
Adapted from https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_int.ml


src/Core__Int.resi
```diff
@@ -0,0 +1,273 @@
...
+`v`. `~radix` specifies the radix base to use for the formatted number.
+
+## Examples
+
+```rescript
+Int.fromString("0") == Some(0)
+Int.fromString("NaN") == None
+```
+*/
+let fromString: (~radix: int=?, 'a) => option<int>
```
aspeddro, line 262: This function signature should not be `(string, ~radix: int=?) => option<int>`?

src/Core__Int.resi
```diff
@@ -0,0 +1,274 @@
...
+
+```rescript
+Int.toFloat(100) == 100.
+Int.toFloat(2) == 2.0
+```
+*/
+external toFloat: int => float = "%identity"
+
+/**
+`fromFloat(n)` return an `int` representing the given value.
```
zth, line 241: Maybe add a short sentence on how it rounds the underlying float? And add an example or two for that. `Int.fromFloat(1.5)` and `Int.fromFloat(1.3)` for example.

src/Core__Int.resi
zth: Yeah that seems more logical. Although we should preserve the current order of the arguments.

src/Core__Int.resi
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+`fromString(~radix?, v)` return an `option<int>` representing the given value
+`v`. `~radix` specifies the radix base to use for the formatted number.
+
+## Examples
+
+```rescript
+Int.fromString("0") == Some(0)
+Int.fromString("NaN") == None
```
zth, line 260: Maybe add an example using another radix as well?

zth (commented):
Looking great! 😄 Just a few minor comments.

zth (approved):

PR #41, opened 2023-02-16
Docs for Promise

aspeddro:
- Add examples header
- Add MDN links
- Add module docstring

### DIFF
--- a/src/Core__Promise.resi
+++ b/src/Core__Promise.resi
@@ -6,10 +6,19 @@
 //
 // More details about polymorphism / invariance / covariance,... can be found here:
 // https://caml.inria.fr/pub/docs/manual-ocaml/polymorphism.html#ss:variance:abstract-data-types
+
+/***
+Functions for interacting with JavaScript Promise.
+See: [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
+*/
+
 type t<+'a> = promise<'a>
 
 /**
-[resolve(value)] creates a resolved Promise with a given `value`
+`resolve(value)` creates a resolved Promise with a given `value`.
+See [`Promise.resolve`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) on MDN.
+
+## Examples
 
 ```rescript
 let p = Promise.resolve(5) // promise<int>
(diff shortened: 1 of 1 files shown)


zth:
Is this ready for review?

src/Core__Promise.resi
```diff
@@ -194,44 +251,46 @@ all(promises)
...
 external all6: ((t<'a>, t<'b>, t<'c>, t<'d>, t<'e>, t<'f>)) => t<('a, 'b, 'c, 'd, 'e, 'f)> = "all"
 
 /**
-`done` is a safe way to ignore a promise. If a value is anything else than a promise, it will raise a type error.
+`done(p)` is a safe way to ignore a promise. If a value is anything else than a
+promise, it will raise a type error.
 */
 external done: promise<'a> => unit = "%ignore"
 
+// TODO: add docstring
```
cristianoc, line 294: These are reserved for the compiler. Can be removed from here as not part of the public api.
Assuming removing them does not stop async code from compiling after opening Core: one just needs to check this. I think it's OK as the Js.* namespace is still visible.

zth: We can try again, but IIRC the reason I added them was because async/await didn't work otherwise. Maybe that's something that can be fixed though.

cristianoc: Can you check again? And if so add an issue to the compiler.

zth: Yes, I'll do that

zth: It seems to work just fine without them. Interesting, I wonder what errored back when I decided to add them... Anyway, we can go ahead and just remove them from `Core`. I'll do that directly in main.

aspeddro:
Ready for review

src/Core__Promise.resi
```diff
@@ -19,13 +28,27 @@ let p = Promise.resolve(5) // promise<int>
...
 
+/**
+`reject(exn)` reject a Promise.
+See [`Promise.reject`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) on MDN.
+
+## Examples
+
+```rescript
+let p = Promise.make((_, reject) => {
+  reject(. #error)
```
zth, line 39: Maybe change this to use a custom defined exception, just to make it clear it's indeed an exception and that you can do that?

src/Core__Promise.resi
```diff
@@ -167,12 +201,35 @@ resolve(5)
...
+  Promise.make((resolve, _) => {
+    Global.setTimeout(() => {
+      resolve(. name)
+    }, ms)->ignore
+  })
+}
+
+let promises = [racer(1000, "Turtle"), racer(500, "Hare"), racer(100, "Eagle")]
+
+race(promises)-then(winner => {
```
zth, line 221:
```suggestion
race(promises)->then(winner => {
```

zth (approved):
Just 2 small things and then we're good to go. Thank you!

PR #44, opened 2023-02-17
Docs for Math

aspeddro:
Adapted from https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_math.ml

### DIFF
--- a/src/Core__Math.resi
+++ b/src/Core__Math.resi
@@ -257,6 +257,19 @@ module Int: {
   @val external pow: (int, ~exp: int) => int = "Math.pow"
 
   // NOTE: In JS `Math.sin(2)` return a float
+
+  /**
+  `sign(a)` returns the sign of its integer argument: `-1` if negative, `0` if
+  zero, `1` if positive. See [`Math.sign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) on MDN.
+
+  ## Examples
+
+  ```rescript
+  Math.Int.sign(3) // 1
+  Math.Int.sign(-3) // 1
+  Math.Int.sign(0) // 0
+  ```
+  */
   @val external sign: int => int = "Math.sign"
 }
 @val external abs: float => float = "Math.abs"
--- a/src/Core__Math.resi
+++ b/src/Core__Math.resi
@@ -172,7 +172,8 @@ module Int: {
   /**
   `imul(a, b)` returns 32-bit integer multiplication. Use this only when you
   need to optimize performance of multiplication of numbers stored as 32-bit
-  integers. See [`Math.imul`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/imul) on MDN.
+  integers.
+  See [`Math.imul`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/imul) on MDN.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


src/Core__Math.resi
```diff
@@ -0,0 +1,297 @@
...
+
+  ## Examples
+
+  ```rescript
+  Math.Int.maxMany([1, 2]) // 2
+  Math.Int.maxMany([-1, -2]) // -1
+  Math.Int.maxMany([]) == 0 // false
+  ```
+  */
+  @variadic @val external maxMany: array<int> => int = "Math.max"
```
aspeddro, line 242: See my notes.

Should we remove it?

aspeddro: This function is [deprecated](https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_math.ml#L499)

src/Core__Math.resi
```diff
@@ -0,0 +1,297 @@
...
+
+  // NOTE: the return type may be not int, example `pow(3, ~exp=-3)` return a float
+  // We should remove this function?
+
+  /**
+  `pow(a, ~exp)` raises the given base `a` to the given exponent `exp`. See [`Math.pow`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/pow) on MDN.
```
aspeddro, line 248: This function is [depreacted](https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_math.ml#L499)

We should remove it?

src/Core__Math.resi
```diff
@@ -0,0 +1,297 @@
...
+
+  ```rescript
+  Math.Int.pow(2, ~exp=4) // 16
+  Math.Int.pow(3, ~exp=4) // 81
+  ```
+  */
+  @val external pow: (int, ~exp: int) => int = "Math.pow"
+
+  // NOTE: In JS `Math.sin(2)` return a float
+  @val external sign: int => int = "Math.sign"
```
aspeddro, line 260: `sign` function signature should be `float => float`, since `Math.sign` return a float?

[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sin)

src/Core__Math.resi
```diff
@@ -0,0 +1,807 @@
...
+  Math.Int.max(1, 2) // 2
+  Math.Int.max(-1, -2) // -1
+  ```
+  */
+  @val external max: (int, int) => int = "Math.max"
+
+  // NOTE: `Math.Int.maxMany([])` compile to `Math.max()` and return `Infinity`
+  // In JS `Math.max([])` return 0
+  // We should create a wrapper to handle an empty array?
+  // Same problem with `Math.Int.minMany`
```
zth, line 231: Let's open a separate GitHub issue for this perhaps, what do you think? Same for the `pow` stuff below.

zth (approved):
Great stuff! Great work @aspeddro .

src/Core__Math.resi
```diff
@@ -0,0 +1,807 @@
...
+  ```
+  */
+  @val
+  external sqrt2: float = "Math.SQRT2"
+}
+
+/**
+Provide Math utilities for `int`
+*/
+module Int: {
```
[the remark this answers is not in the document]
aspeddro: This suggestion does not show docstring on hover.

src/Core__Math.resi
aspeddro: Ok

src/Core__Math.resi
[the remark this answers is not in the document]
zth: Right, that's a bug we currently have. Module level docstrings in signatures aren't picked up for some reason. Does the docstring you had previously show up? If so we can keep it and then do another pass moving things to proper module level docstrings when we've fixed that bug.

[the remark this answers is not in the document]
aspeddro: Yes.

image

src/Core__Math.resi
aspeddro: See #53

src/Core__Math.resi
aspeddro: See #53

PR #46, opened 2023-02-17
docs for Error

zth:


src/Core__Error.resi
```diff
@@ -0,0 +1,157 @@
+/***
+Functions for working with JavaScript exceptions.
+
+See [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) on MDN.
+*/
+
+/** Represents a JavaScript exception. */
+type t = Js.Exn.t
+
+external fromException: exn => option<t> = "?as_js_exn"
```
zth, line 10: Leaving this right now because I quite frankly don't fully understand why it does, or rather when I'd use it.

cristianoc (commented):
I think this checks if an exception is a js exception (as opposed to a ReScript one)

cknitt (approved):

PR #48, opened 2023-02-18
Add array.at function

dkirchhof:


src/Core__Array.resi
```diff
@@ -103,3 +103,21 @@ let filterMap: (array<'a>, 'a => option<'b>) => array<'b>
 let shuffle: array<'a> => array<'a>
 let shuffleInPlace: array<'a> => unit
 let flatMap: (array<'a>, 'a => array<'b>) => array<'b>
+
+
+/** 
+  `at(array, index)`
+
+  Get an element by it's index. Negative indices count backwards from the last item.
```
glennsl, line 111:
```suggestion
  Get an element by its index. Negative indices count backwards from the last item.
```

Apostrophe can be used to indicate either possession or contraction, and while in this case you do mean possession there's an exception to the rule for pronouns because it conflicts with contraction. "it's" means either "it is" or "it has", and you should therefore drop the apostrophe to indicate possession.

(Yes, the English language is crazy complicated :roll_eyes: )

dkirchhof: You're completely right.

zth (approved):
Nice!

zth:
Oh, @dkirchhof , would you also add an entry to the changelog about this being added?

zth (approved):

PR #50, opened 2023-02-18
Fix incorrect type signatures in Array

glennsl:
Unless I'm misunderstanding something, these type signatures shouldn't be unsoundly typed, that is, the return type shouldn't be independent of the input argument types.


zth (approved):
Good catch, thank you! This looks like an oversight. Would you add an entry to the changelog too?

glennsl:
Can do. I'm unsure of the format of the changelog though. Should I just put this under the `API Changes` heading?

```
- Fixed type signatures of `Array.fromArrayLikeWithMap` and `Array.fromIteratorWithMap`. https://github.com/rescript-association/rescript-core/pull/50
```

zth:
Yes, that'll be good.

zth (approved):

PR #54, opened 2023-02-18
Docs for Float

aspeddro:


src/Core__Float.resi
```diff
@@ -1,26 +1,404 @@
...
+Float.isFinite(1.0) // true
+Float.isFinite(Float.Constants.nan) // false
+Float.isFinite(Float.Constants.positiveInfinity) // false
+```
+*/
+@val
+external isFinite: float => bool = "isFinite"
+
+/**
+`parseFloat(v)` parse the given `v` and returns a float. Leading whitespace in
```
zth, line 142: I think we should add a note here that Float.fromString can be used if you want to ensure you get a valid float back and not NaN.

src/Core__Float.resi
```diff
@@ -1,26 +1,404 @@
...
+Float.parseFloat(3.0) // 3.0
+Float.parseFloat("3.14some non-digit characters") // 3.14
+Float.parseFloat("error")->Float.isNaN // true
+```
+*/
+@val
+external parseFloat: 'a => float = "parseFloat"
+
+/**
+`parseInt(v)` parse the given `v` and returns a float. Leading whitespace in
```
zth, line 160: I wonder if we should keep this + other parseInt stuff around. Doesn't make sense to have it in the Float module imo, unless I'm missing something.

src/Core__Float.resi
```diff
@@ -1,26 +1,404 @@
...
+/**
+`fromString(~radix?, str)` return an `option<int>` representing the given value
+`str`.
+
+## Examples
+
+```rescript
+Float.fromString("0") == Some(0.0)
+Float.fromString("NaN") == None
+Float.fromString("6") == None
```
zth, line 366: Why is this None?

src/Core__Float.resi
```diff
@@ -1,26 +1,404 @@
...
+*/
 let fromString: 'a => option<float>
+
+/**
+`toInt(v)` returns an int to given float `v`.
+
+## Examples
+
+```rescript
+Float.toInt(2.0) == 2
```
zth, line 377: Might be interesting to have an example for what happens when you pass something with a fraction, like 1.1 or 1.6.

zth (commented):
Great stuff! A few comments.

src/Core__Float.resi
glennsl: Both should also explain what the expected format is. Consumers might expect it to be locale-aware, or to be able to parse rescript float literals. The expected format is that of JS number literals.

src/Core__Float.resi
aspeddro: Fixed c49fcdf

src/Core__Float.resi
aspeddro: Fixed c49fcdf

src/Core__Float.resi
```diff
@@ -1,26 +1,406 @@
...
+
+## Examples
+
+```rescript
+Float.fromString("0") == Some(0.0)
+Float.fromString("NaN") == None
+Float.fromString("6") == Some(6.0)
+```
+*/
 let fromString: 'a => option<float>
```
aspeddro, line 369: it shouldn't be `string => option<float>`?

zth: Yes, no idea why it's not. You can go ahead and change it (and add an entry to the changelog please).

src/Core__Float.resi
aspeddro: > I think we should add a note here that Float.fromString can be used if you want to ensure you get a valid float back and not NaN.

For this, we must change the signature of `parseFloat` to `string => float`

The signature is currently `'a => float`

src/Core__Float.resi
aspeddro: We can do this in another PR?

zth: Yes, for sure.

src/Core__Float.resi
aspeddro: Done https://github.com/rescript-association/rescript-core/pull/54/commits/e003d749ab51126a90be31200398296115a2349f

zth:
@aspeddro this is ready to go, right?

aspeddro:
> @aspeddro this is ready to go, right?

yes

PR #56, opened 2023-02-19
Flesh out Console

### LINKED ISSUE #55: Console bindings incomplete
`console` has a lot more functions than the currently bound ones, see https://developer.mozilla.org/en-US/docs/Web/API/console.

glennsl:
Fixes #55

### DIFF
--- a/src/Core__Array.resi
+++ b/src/Core__Array.resi
@@ -104,7 +104,6 @@ let shuffle: array<'a> => array<'a>
 let shuffleInPlace: array<'a> => unit
 let flatMap: (array<'a>, 'a => array<'b>) => array<'b>
 
-
 /** 
   `at(array, index)`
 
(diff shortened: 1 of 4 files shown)


zth (commented):
This is great, thank you! Can we revert the unrelated formatting changes in Array and Math?

glennsl:
Done. Not sure why Github doesn't update, but the commits have been removed from the branch that the PR points to.

It's going to really annoy me if the formatting isn't fixed elsewhere though, since I'll constantly have thee uncommitted formatting changes hanging around. It seems like a good idea to set up CI workflow to check formatting, so as to avoid this in the future.

zth:
Sure, I don't mind that. Would you be up for PR:ing CI and a command in package.json/scripts for formatting everything?

PR #58, opened 2023-02-19
chore(ci): add style check workflow

glennsl:


glennsl:
Tested that it fails when it should on my own fork. See https://github.com/glennsl/rescript-core/actions

zth:
Cool! Can we add a description somewhere that shows up in GitHub that you'll need to format all files (and point to the command to do that)? Just do people understand what's wrong when this fails.

glennsl:
Done. See https://github.com/glennsl/rescript-core/actions/runs/4217814412 for an example.

zth:
Fantastic, thank you very much!

PR #60, opened 2023-02-19
chore(test): fix undefined __dirname in node 19

glennsl:
`__dirname` is apparently in es modules as of node 19. This fixes that by using `import.meta.url` instead.

Also adds current and latest LTS node versions (19 and 18 respectively) to the CI matrix.


zth (approved):
Good catch.

PR #62, opened 2023-02-20
fix: remove Date.setDay and Date.setUTCDay as these functions doesn't exist

dkirchhof:
While writing the docs I stumbled across these functions. They don't exist in js and throw errors.


zth:
Nice find! They should be setDate, right?

zth:
Haha 😉  Great find, thank you! Merging.

PR #68, opened 2023-02-23
docs(option): fixes for several docstrings

glennsl:
I found the existing docstrings confusing, so I propose some simplifying changes.

### DIFF
--- a/src/Core__Option.resi
+++ b/src/Core__Option.resi
@@ -66,7 +66,7 @@ Option.forEach(None, x => Console.log(x)) // returns ()
 let forEach: (option<'a>, 'a => unit) => unit
 
 /**
-`getExn(opt)` raises an Error in case `None` is provided.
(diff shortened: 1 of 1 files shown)


zth (approved):

src/Core__Option.resi
```diff
@@ -96,8 +96,7 @@ Option.getUnsafe(None) // Raises an error
 external getUnsafe: option<'a> => 'a = "%identity"
 
 /**
-`mapWithDefault(opt, default, f)` applies `f` to `opt`, if `opt` is `None`, then
-`f` return `default`, otherwise return that value applied with `f`.
+`mapWithDefault(opt, default, f)` returns `Some(f(value))` if `opt` is `Some(value)`, otherwise `default`.
```
cknitt, line 99: Shouldn't this be "... returns f(value)"?

glennsl: Ah, yeah it should be. Good catch! Fixed in https://github.com/rescript-association/rescript-core/pull/70

PR #81, opened 2023-03-02
Remove Math.Int.pow

aspeddro:
This function has been [deprecated](https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_math.ml#L499) for int. The return type may be not int, example `Math.Int.pow(3, ~exp=-3)` return a float


glennsl:
Could we provide one that does provide this but isn't a direct binding? I think that would be quite useful.

Issue #77, opened 2023-02-28
compare and equal


cknitt:
Currently, we have `cmp` and `eq` for the `List`, `Option` and `Result` modules (coming from Belt).

It would be good to have these available for more modules, e.g. `Array`, `Int`, `Float`, `String`, `Date`, `Null` or `Nullable`.

Naming-wise, maybe `compare` and `equal` would be more "ReScript-like".

zth:
Agreed, both on adding them in more places, and about the naming 👍

cknitt:
Great, I can create a PR for that in the next days.

jmagaram:
I was just about to ask for this!

cknitt:
Please have a look at the above draft PR.

What other modules would you like to see `compare` and `equal` for?

jmagaram:
I have some thoughts about this. Will post today.

jmagaram:
There are a few issues:

- `cmp` functions, those like `('a,'a)=>int`. Should we have built-in versions for `int`, `float`, etc.?
- utilities that construct/map `cmp` functions
- utilities that make practical use of `cmp`, like `equals`, `lessThan`, `max`

My original gripe was that there are lots of times I need a `cmp` - sorting an array or making a Belt.Map - and I find myself writing an inline ternary like `[3,2,7]->Array.sort((x,y)=>x<y ? -1 : x>y ? 1 : 0)`. Annoying. **Is this what everyone does?** It would be convenient if I could do `[3,2,7]->Array.sort(Int.cmp)`. So maybe add these `cmp` functions to the types it makes sense like `int`, `float`, `date`, `bigInt`. `bool` wouldn't hurt and good for consistency. `string` is tricky because there are many ways to compare. I didn't realize Javascript has a way of comparing strings case insensitive, ignoring accents, numeric like "1" < "10". but we could have a default String.cmp which is like `string.localeCompare` (which returns a float by the way) but with no options - a quick and dirty way to sort some strings in a reasonable way. 

But now that we have `cmp` for built-in primitives, what if you want to do a reverse sort? We could have `Int.cmp` and `Int.cmpReverse`? I think if we're going to have one we should have both. 

`Belt.Option.cmp` seems a little weird to me. If you want to build a `cmp` to compare two options I think you'd want a function that takes a `('a, 'a) => int` as input, and generates a `(option<'a>, option<'a>) => int` as output. And then feed that result into Array.sort or something. I don't know how you'd use the function as it is designed now.

[...]

PR #84, opened 2023-03-03
compare and equal

cknitt:
Add compare and equal functions, see #77.

Modules:

- [x] Array
- [x] Int
- [x] Float
- [x] String
- [x] Date
- [x] Null
- [x] Nullable

TODO:

- [ ] Docs
- [ ] Tests


jmagaram:
I put some comments in the issue. I think we should discuss there, not here. I see problems using `compare` with options when sorting arrays - data first is awkward. Same with instantiating a Belt.Set or Map. I'm not sure where data first really works for option since you're going to end up with a number like -1, 0, or 1 and then what to do with it? Also does the subtraction implementation for int.compare work with NaN and other weird cases; it might.
```
let compareInt = (x, y) => x < y ? -1 : x > y ? 1 : 0
let x = [Some(1), None, Some(5), Some(2)]->Array.sortInPlace((a, b) => Option.cmp(a, b, compareInt))
```

jmagaram:
The implementation fails with overflow like `Int.Constants.minValue - Int.Constants.maxValue `. I'd recommend this. Make sure to explicitly specify the types or else you'll end up with `Call_obj.lessthan`.
```
let compare = (x: int, y: int) => x < y ? -1 : x > y ? 1 : 0
```

jmagaram:
Might want to explain in the docs which equality operator is used for floats/ints. if you use === then NaN != NaN and -0.0==+0.0 which seems fine to me. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

jmagaram:
I'm confused about the string compare flavors. How does the version you've coded relate to `String.localeCompare` with or without options? Is there any advantage to it? The docs should explain when to use which ones. I'm wondering if these compare options should be grouped/unified together so the developer can easily see their choices together. Maybe `String.compareLocale`, `String.compareLocaleWith`, and `String.compare`, where the last is the version you've coded and the first two are the `String.localeCompare` versions with and without options.

https://262.ecma-international.org/7.0/#sec-abstract-relational-comparison

cristianoc (approved):

PR #88, opened 2023-03-05
switch global null to come from Nullable module and not Null

zth:
This changes the globally available `null` value to originate from `Nullable` rather than `Null`. Makes sense imo given that global `null` and `undefined` now originate from the same module.

It also makes working with domRefs in React more natural, allowing you to write `null` directly instead of `Nullable.null`:
```rescript
@react.component
let make = () => {
  let divRef = React.useRef(null)
  <div ref={divRef->ReactDOM.Ref.domRef}> {"Hello"->React.string} </div>
}
```

(side note - would be cool to get completion for `ReactDOM.Ref.*` in the scenario below. Might be able to work that into the editor tooling)

Thoughts? cc @cknitt @glennsl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Remove `reduceReverse` in favor of `reduceRight`. https://github.com/rescript-association/rescript-core/pull/49
 - Fixed type signatures of `reduce` and `reduceWithIndex`. https://github.com/rescript-association/rescript-core/pull/49
 - Add `panic`/`Error.panic`. https://github.com/rescript-association/rescript-core/pull/72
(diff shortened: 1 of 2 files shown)


cknitt:
I like it! 🙂

glennsl:
Ditto! :)

glennsl (approved):

cristianoc:
What about typical bindings, is the idea that one most often encounters null | undefined | T, rather than null | T?
Asking as the few binding I looked at to add genType support to Core were using null | T.
Otherwise one has o bind with `Null.t` and will have most often to reach for `Null.null`.

glennsl:
Good question. I seems to recall the decision was made long ago to prefer `Nullable.t` over `Null.t` and `Undefined.t`, just for simplicity. I don't think React `ref`s have ever been set to `undefined`, so I believe that is the reason why it's typed as it is.

I wonder if it would be practical to make `Null.t` a subtype of `Nullable.t`...

PR #92, opened 2023-03-06
array of result to result of array

jmagaram:
convert array of result to result of array


jmagaram:
I have tests that all pass but I couldn't figure out how to run them in your system. Tests like this. They were pretty trivial to write.
```
  test("when only error, take first", _ => {
    let input = [Error("a"), Error("b")]
    let res = input->ResultUtil.fromArray
    res->expect->toEqual(Error("a"))
  })
```

jmagaram:
I have had to write this function MANY times. Writing it in functional style is a little tricky, requiring `array.reduce`. I finally tried to do it in imperative style for better performance, and here it is.

glennsl:
There's no reason a recursive version wouldn't be just as fast. There's also a couple other things we can do to make it significantly faster, 1. create the whole array uninitialized, and 2. use unchecked array get and set.

Haven't tested this for either performance or correctness, but have a look at [the generated JavaScript](https://rescript-lang.org/try?version=v10.1.2&code=DYUwLgBAZgTg9gWwIIxgQwJ4HUCWYAWEAvBABQAeAzgDTQCUxAfBAN4BQEEokIqcMxCDBBRSAOTgA7EHQ5dwEHJIAmIcoOGiADLM7cIcANaVBAbQC6c-QCNhaQxpGkoaYJRlyA7vhygIAQlsQewA6AGMpMBBJMBN2Tk5KTzwwwioAWkYUdAwQgHNwUiVVcnDI6NiGeISAHwgJaSYIIPsIAC4SMBgAVxA5TjqAZUQQCgYiRn6EpJTCUXIqqYGIAHlDUiNxyYSd+UgAfUEjSkyAKUoQ7MwQgAduynwNw11dzmK1dpJ30oiYipMANQQACMSwgdQAonwYKQZE1qq9OLx4AIOhBhghRh5EQkWg40V1emDOABfMFknYUiBUmZgVIQZH8Mp-GJxOR1BogJprJ6UF5DEaw1BbCBQlFCmAvMlktgAAWkngZ5CiMEkrggCHsIAAqpIlHgcK4cAAvEDKXWUNBQEBtRQxJpoVCYAA8AHI0MwSAAiK4YL1sNQqtXACAFMAWq02siOnJuj20JRgEXuwRegCkMcw+26kkt1v2Yf9gd4wYg7nDucjttImYwccYCZitHdIpzeFTGadGGzlfz5f9bH0sEQvtwBAATIIKDR6PCrApjoJNYYdXq9WBDcATWaI9bp5lfSFQJI8gRZPPIMIwlw4HAboIcEwpjgoIoIIwSBksl2j9FT4QEQSHljheUkGTcLlAMSZI6TSE5GDDXdRhwOhMlfKDwVWdYFifHFjkycskKKWgFmJG87yKCAgWBUDalFaFSH2Bg0BMZEmmRckphlPRbxuUgdDYMkgA). It might surprise you :)

```rescript
@new external makeUninitializedUnsafe: int => array<'a> = "Array"

let fromArrayWith2 = (xs, f) => {
  let oks = makeUninitializedUnsafe(xs->Array.length)

  let rec loop = i =>
    if i >= xs->Array.length {
      Ok(oks)
    } else {
      switch xs->Array.getUnsafe(i)->f {
      | Ok(x) =>
        oks->Array.setUnsafe(i, x)
        loop(i + 1)
      | Error(_) as err => err
      }
    }

  loop(0)
}
```

glennsl:
> I have tests that all pass but I couldn't figure out how to run them in your system. Tests like this. They were pretty trivial to write.
> 
> ```
>   test("when only error, take first", _ => {
>     let input = [Error("a"), Error("b")]
>     let res = input->ResultUtil.fromArray
>     res->expect->toEqual(Error("a"))
>   })
> ```

This is a translation of that test to the framework used here (given that `eq` is defined in the test module):

```rescript
Test.run(__POS_OF__("when only error, take first"), [Error("a"), Error("b")]->Result.fromArray, eq, Error("a"))
```

jmagaram:
Ok trying to incorporate tests. Also not compiling because I was using a different Array module that had a `push` method. Confusing. Will work it out.

jmagaram:
I'm not too savvy with pull requests. My last two commits are for `Result.mapError`, a useful function for railway-oriented programming and getting all the errors in a consistent form. F# has this in their core library.

jmagaram:
Do lists get used? For completeness maybe there should be a version for lists? What's a good name for this? Where will it get discovered?

Result.fromArray
Result.fromList

Result.fromArrayOfResults
Result.fromListOfResults

Result.collectFromArray
Result.collectFromList

Result.fromResultArray
Result.fromResultList

Result.fromManyInArray
Result.fromManyInList

And/Or flip side...

Array.toResult
Array.collectResults
Array.reduceResults

cknitt:
I also find `mapError` and `fromArray` useful. 👍 Not sure about the naming for `fromArrayWith`. With what? 🙂 Maybe it should be `fromArrayWithMap`? As we also have `Array.fromWithMap`.

I would not add more functions for now.

jmagaram:
I feel very confident that `Result.mapError` should be included. This is in almost every other result module - F#, Rust, fp-ts.

I'm less confident in the `fromArrayWith` and `fromArray`. In the functional programming world these are `traverse` and `sequence`. fp-ts has a `traverseWithIndex`. The F# article below shows a good example of transforming an option, which you can think of as a list of 0 or 1 items, into a result. There are flavors of this kind of function where you get all the errors, which seems useful but less useful. F# doesn't have this. I think the terms "sequence" and "traverse" are a bit technical and prefer terms like `fromArrayWithMap` or `fromMany` or `fromManyWithMap` or `invertArrayWithMap` etc.
```
let fromOption: (option<'a>, 'a => result<'b, 'c>) => result<option<'b>, 'c>
```
https://fsharpforfunandprofit.com/posts/elevated-world-4/#traverse
https://gcanti.github.io/fp-ts/modules/Traversable.ts.html
https://stackoverflow.com/questions/50789065/turn-list-of-result-into-result-of-list-inside-a-computation-expression

zth:
I agree `mapError` looks useful and within scope, but `fromArray` doesn't feel broad enough. Would you split these up in separate PRs so it's easier to treat them individually?

jmagaram:
Closing since I split this into two others.

Issue #86, opened 2023-03-04
On number parsing


glennsl:
This touches a bit on the discussion in #83, but covers more broadly the current options for parsing numbers in JavaScript, to try to inform which semantics we should expose, more so than just how we should type and organize the JS API.

Currently, I believe the only option for parsing numbers are `parseInt`/`parseFloat`. Either directly through `Float.parseFloat` and `Float.parseInt`, but also as the underlying API used in `Float.fromString` and`Int.fromString`. While there unfortunately aren't any good options, in my opinion `parseInt` and `parseFloat` is the worst of all the bad options. For the sake of easy comparison and to facilitate efficient discussion, I've outlined the pros and cons of the options i know of below.

Note that I've only covered the differences between the options. Common peculiarities to all of them is that:
- They accept scientific notation, but not optionally so it can't be turned off.
- They don't distinguish parse failure from the actual valid value of `NaN`. 
- They don't support group separators.
- They don't provide locale-aware parsing.

## `parseInt`/`parseFloat`

The core of the prom with these functions is described by the following quote [from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat#description):

> parseFloat() picks the longest substring starting from the beginning that generates a valid number literal. If it encounters an invalid character, it returns the number represented up to that point, ignoring the invalid character and all characters following it.

While this may not seems like such a big issue, since it just ignores certain kinds of "mistakes", keep in mind that there isn't just one number format used across the world. These functions will only parse a very simple number format, close to but not entirely the same as JavaScript number literals. And iif it ecnounters anything that doesn't fit that, it will just ignore the rest. For example, if any kind of group separator is used, that and everything that follows will be silently ignored. The same goes for using a decimal separator other than `.`.

In short, these invocations will all return `15`:

```js
parseInt("15,123");
parseInt("15 123");
parseInt("15 * 3");
parseInt("15px");
```

### Pros
- Simple functions that are easy to bind to.
- Zero-cost.
- Accepts optional radix argument.

### Cons
- Will accept any string that starts with something that can be parsed as a number.

## Number coercion

There are a number of ways to trigger number coercion, such as using any numeric operator on them. E.g. if `value` is a string, `+value` will return a number.

This improves on `parseInt` and `parseFloat` most notably by rejecting input that isn't wholly parsable, such as all the examples in the section above. It does come with a few extra cons though, but they are for the most part possible to work around.

### Pros
[...]

PR #93, opened 2023-03-07
Stricter number parsing in Float.fromString and Int.fromString

glennsl:
Implements the changes proposed in #86. See the differences in the tests for specifics on behavioural changes.

Note also that this drops the `radix` option of `Int.toString`, but on the other hand improves parsing of radix prefixes in the string itself, i.e. `0x`, `0o` and `0b`.


jmagaram:
Got a big laugh reading this. Especially “worst of all the bad options”. And interesting to learn there is a Number object distinct from the primitive - glad to have that in my toolbox. A couple thoughts…

Group them together so the user will see a menu of related choices. So parseString not fromString. If you want to encourage use of your implementation call it `parse`. 

In the doc help for each explain the limitations and behavior like “Unreliable behavior if the string includes separators or characters not part of the number.” Give a few bad examples. Point to MDN.

glennsl:
Superseded by https://github.com/rescript-lang/rescript/pull/8129

PR #98, opened 2023-03-13
Result.mapError

jmagaram:
This is a serious omission. Almost every functional library with Result or Either types have this. If you want to do railway-oriented programming you often need to convert the error types to a common type, and this function does it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next version
 
+### API changes
+
+- Add `Result.mapError` https://github.com/rescript-association/rescript-core/pull/98
+
 ## 0.4.0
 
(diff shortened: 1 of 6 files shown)


tsnobip:
hey, why don't we merge this? It'd definitely be welcome!

DZakh:
Agree. Actually, quite some time ago I wrote in the forum about adding the `Result.mapError` and people there accepted the idea. Also, I read the source code of many libraries and almost all of them include the handwritten `mapError` function. I think there's no reason for it not to be a part of the `Core` lib. Ping @zth 
https://forum.rescript-lang.org/t/suggestion-to-add-operations-on-result-type/3070

PR #103, opened 2023-03-16
Object.freeze and Object.isfrozen documentation

jmagaram:
Modified the object but does not return `unit`


jmagaram:

PR #104, opened 2023-03-17
Object.seal/isSealed documentation

jmagaram:
Stole liberally from MDN


jmagaram:
Modifies the object but does not return `unit`. The code examples are very weird. With typical Rescript Object syntax everything is immutable. But if you use Object.set all bets are off, the compiler doesn't know the shape of your objects, etc. Hard to teach this in a tiny doc string.

jmagaram:

PR #105, opened 2023-03-17
Object.preventExtensions/isExtensible documentation

jmagaram:
Note this function modifies the object but does not return `unit`.


jmagaram:

PR #106, opened 2023-03-17
Remove deprecated j` string interpolation

cknitt:
Get rid of "j" string interpolation that is deprecated in 10.1.4 and removed in 11.0.
(Otherwise Core will not build with the latest master of the compiler.)


cknitt:
Duplicate of #107.

PR #110, opened 2023-03-17
Array.fromOption

jmagaram:
Includes full documentation and tests. Under the hood uses Array.of to generate an empty array.


jmagaram:
Making this a separate PR with only this function, will close this and do it again

PR #115, opened 2023-03-20
Object.assign documentation and tests

jmagaram:
This took me a while because I wrote some tests to experiment and got the source and target mixed up - the mutable target goes first. Should we have labels on these?

I attempted to write code for `assignMany` but it doesn't work because you can't create an array of objects with different types. Well you can I guess if they start off as Object.empty() or are imported - jumping through hoops. Will this really get used? Do we need a `toGenericObject` function?

I couldnt' figure out the other flavor of assign with the json thing and what it does.


zth (approved):
Looking great!

Good point about `assignMany`. I think we can leave it undocumented for now. The `Object` module is for more specialized usage anyway.

glennsl:
But you can also just chain calls to `assign`, so... :shrug:

jmagaram:
Interesting idea on the `assign2` `assign3` - hadn't thought about that. I don't know when any of these functions will get used from ReScript, if ever. I wouldn't mind removing the `assignMany` and rely on chaining calls. A `toUntypedObject` could normalize objects in a list so `assignMany` will work. You all have much more real-world experience and know better what is needed here.

jmagaram:

PR #116, opened 2023-03-20
Result.forEach with tests and docs

jmagaram:
Analogous to Array.forEach and Option.forEach. In nearly every other standard library of Result/Either types.


glennsl (approved):

zth:
Oh, one final thing @jmagaram - please add an entry to the changelog. Then this is mergeable.

jmagaram:
> Oh, one final thing @jmagaram - please add an entry to the changelog. Then this is mergeable.

Sorry don't know how to do that. Is this 0.2.0? Where does it go in that doc?

zth:
Just add a new "API changes" under "main", like now exists under "0.2.0", and add your entry there.

jmagaram:
I'll try to collect the object documentation PRs but I'm expecting to be paid separately for each one! Actually I don't know how to combine them. Will try to learn.

PR #119, opened 2023-03-23
Docs for all specific instances of typed array

jmagaram:
This is for the BigInt64Array. Once people comment and are satisfied almost everything can be cloned to the other specific instances. Few specific things:

- I renamed `make` to `fromArray`
- I put a lot of warnings about unsafe access to buffers. From my limited testing it seems you can screw yourself if the buffer isn't big enough or you go outside the bounds.
- I think `fromLength` is length of BigInts not a byte length but am not totally sure. If so maybe we should use a labeled argument to make it explicit. Someone please confirm.
- I added the `fromIterableWithMap`.


jmagaram:
Quite frustrating. I tried writing some tests to see how this stuff works and wasted a lot of time because BigInt==BigInt is broken in the compiler. https://github.com/rescript-lang/rescript-compiler/issues/6096

zth:
This is looking solid to me 👍 No comments, I think it's on a good level.

Issue #101, opened 2023-03-16
Confusion about re-throwing an exception


jmagaram:
I'm trying to catch and work with an Error, but don't know for sure what I'm receiving in the exception. If I test using `Error.fromException` and that fails and returns `None` I want to throw it again, but I can't. I need to use the OCaml `raise` function which seems to be available to me. Maybe the type of `Error.raise` needs to be loosened? Do we need an `Error.raiseAny`? 
 
```rescript
let f = () => {
  try {
    Console.log("do things")
  } catch {
  | _ as e =>
    switch e->Error.fromException {
    | Some(x) => Console.log("Handled!")
    | None => Error.raise(e) // Compile error, expects an Error.t
    | None => raise(e) // Works!
    }
  }
}
```

PR #120, opened 2023-03-24
Error.raiseAny

jmagaram:
This is my attempt at fixing the problem I raised in https://github.com/rescript-association/rescript-core/issues/101, which is basically that I am inspecting an exception and then want to re-throw the `exn` (not an `Error.t`). I could not find a way to do that with the `Error` module, which is where I would expect to find it. My proposed fix is makes the throw statement return a `unit`, although every other raise command returns a `'a`. What is correct?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### API changes
 
 - Add `Result.forEach` https://github.com/rescript-association/rescript-core/pull/116
+- Add `Error.raiseAny` https://github.com/rescript-association/rescript-core/pull/120
 
 ## 0.2.0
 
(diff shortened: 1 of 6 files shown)


glennsl (commented):
You're confusing errors and exceptions I think. [Exceptions](https://rescript-lang.org/docs/manual/latest/exception) are a language-level concept, and it comes with its own `raise` function that will do what you want here.

Errors, as defined here, refers to JS `Error` objects. It doesn't make sense to have a function here specifically for raising anything that's not a JS Error object.

jmagaram:
Ok. I think because I was working with an `Error.t` and my mind was in that place, I expected to find what I was looking for there. And then when I looked at the `raise` function, I saw it was in `Pervasives` which I think is something this library was trying to clean up. `Pervasives` has a lot of stuff that does not belong. And on the ReScript doc web site `raise` is not in the Syntax Lookup. Maybe the docs for `Error.raise` should point people to the ReScript docs on Exceptions in general. Someone else can close this.

PR #123, opened 2023-03-24
Error.getUnsafe

jmagaram:
Provides a way to get a custom property value out of an Error. Includes full documentation, warnings, and a test. This satisfies a real-world scenario of me trying to get the "code" property out of a Firebase SDK exception. See discussion in 
https://github.com/rescript-association/rescript-core/issues/102.

### DIFF
--- a/src/Core__Error.resi
+++ b/src/Core__Error.resi
@@ -169,3 +169,27 @@ Error.panic("Uh oh. This was unexpected!")
 ```
 */
 let panic: string => 'a
+
+/**
+`getUnsafe` returns a custom property on an error object, given its name. This is useful for working with custom exceptions thrown from JavaScript.
+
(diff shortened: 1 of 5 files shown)


src/Core__Error.resi
```diff
@@ -169,3 +169,27 @@ Error.panic("Uh oh. This was unexpected!")
...
+switch exn->Error.fromException {
+| None => raise(exn) 
+| Some(err) =>
+  switch err->Error.getUnsafe("code") {
+  | None => raise(exn)
+  | Some(code) => {
+      if (code === "invalid-password") {
+        Console.log("Try again!")  
+      }
+    }
```
glennsl, line 189: I find this more readable and easier to evolve:

```suggestion
  | Some("invalid-password") => Console.log("Try again!") 
  | Some(_) => ()
```

glennsl (commented):
Agree that it would be better to have a beefed-up `Type` module with a generalized version of this.

src/Core__Error.resi
jmagaram: Yes much better. Thanks.

Issue #85, opened 2023-03-03
fill out the Option module?


jmagaram:
I like using Option. I'm wondering if the module should be filled out with some functions that are available in Option packages from other languages. Most of these functions are simple conveniences that aren't strictly necessary. But they lead to shorter and sometimes easier-to-understand code. If we provide conversion functions from both sides, like option->null and null->option, the programmer finds it more easily. Maybe standardize on terminology for lazy forms. These usually take just a few lines to implement. What do you think? Should we consider augmenting the Option module? Similar changes could be made to Result as well. I've looked at [F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html) and [fp-ts.](https://gcanti.github.io/fp-ts/modules/Option.ts.html) Here are some ideas. If there is interest in this, I can do more research and make a proposal.

```
module type Ideas = {
  let valueExn: option<'a> => 'a // same as getExn
  let valueUnsafe: option<'a> => 'a
  let valueOrDefault: (option<'a>, 'a) => 'a
  let valueOrDefaultWith: (option<'a>, unit => 'a) => 'a // new lazy form
  let concat: (option<'a>, option<'a>, ('a, 'a) => 'a) => option<'a>
  let orElseWith: (option<'a>, unit => option<'a>) => option<'a> // lazy form
  let map2: (option<'a>, option<'b>, ('a, 'b) => 'c) => option<'c>
  let map3: (option<'a>, option<'b>, option<'c>, ('a, 'b, 'c) => 'd) => option<'d>
  let fold: (option<'a>, 'b, ('a, 'b) => 'b) => 'b
  let match: (option<'a>, 'b, 'a => 'b) => 'b // same as mapWithDefault?
  let matchWith: (option<'a>, unit => 'b, 'a => 'b) => 'b
  let iter: (option<'a>, 'a => unit) => unit
  let exists: (option<'a>, 'a) => bool
  let contains: (option<'a>, 'a) => bool
  let count: option<'a> => int
  let flatten: option<option<'a>> => option<'a>
  let toArray: option<'a> => array<'a>
  let toList: option<'a> => list<'a>
  let toUndefined: option<'a> => Js.undefined<'a>
  let toNull: option<'a> => Js.null<'a>
  let fromUndefined: Js.undefined<'a> => option<'a>
  let fromNull: Js.null<'a> => option<'a>
  let fromPredicate: ('a, 'a => bool) => option<'a>
}

module Option = {
[...]

PR #124, opened 2023-03-24
Option existential quantifiers

jmagaram:
This adds two functions - `Option.isSomeAnd` and `Option.isNoneOr`. Full and really good documentation and tests. These are convenience functions that lead to shorter and easier-to-understand code. These correspond to the quantifiers "forAll" and "exists" in mathematics. We have analogous functions on the Array module.

`isSomeAnd` is like `Array.some`. Another common name is `exists`. But I like the parallel usage here.

`isNoneOr` is like `Array.every`. Another common name is `forAll`. But I and other people find this naming a little confusing for empty collections. For every PR I've received money for, I got $1 million dollars. Technically true.

Rust discussion on isSomeAnd: https://github.com/rust-lang/rust/issues/93050
F#: https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html


jmagaram:
How about this @zth @aspeddro @DZakh and anyone else? This is a small part of https://github.com/rescript-association/rescript-core/issues/85 and based on analogous functions in Rust and the Array module. As long as we have an Option module I think we should make it more complete and convenient. I am happy to get feedback from the community on a larger set of improvements.

DZakh:
I agree that since we have `Option` module, it should be rich and preferred over `Null`/`Nullable`, but I'm actually skeptical about the PR.

There are a few reasons why I don't like the proposed functions:
- It differs from person to person, but it took me a while to figure out what the functions actually do by their name without taking a look at the code
- There's already `mapOr`, and I think it should be the only function to stay in the `Option` module
  - Since everybody uses the `map` function a lot, `mapOr(false, fn)` is more readable than `isSomeAnd`
  - It uses the parsing/piping approach while `isSomeAnd` and `isNoneOr` is more like validation restricted to returning only `bool`
  - I prefer only one way to do things when DX doesn't suffer. A smaller API is a good thing, especially valuable in bigger teams. That's why golang is popular. And I think making your users [falling into the pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/) is very important.
  
So instead, I'd suggest adding `mapErrorOr` to cover `isNoneOr` use case.

jmagaram:
Thanks for the thoughtful analysis.

As for naming, some alternate names would be `exists(fn)` and `forAll(fn)` or `every(fn)`. I prefer the ones I chose because `forAll` is a bit unintuitive for the empty/none case.

Without `isSomeAnd` first you’d do a `filter` followed by an `isSome`. Without `isNoneOr` you’d do a `filter` followed by `isNone`. Because both of these functions end up using `isX`, I think the proposed functions are easily discoverable in the intellisense menu. When you use `isSome` you notice `isSomeAnd` and realized you could accomplish what you’re trying to do in one API call. 

Alternately you could accomplish both with `mapOr` but I find the resultant code MUCH less intuitive and doesn’t describe my intent very well. This applies to when I am writing it and when I am later looking at the code to figure out what it is doing. This is the key problem these functions solve - making the code match the intent. I think of it as “is x SOMEthing that meets these requirements?” rather than “let me map it to a bool with false for the None case.”

Piping? Don’t understand that argument. If `isSome` and `isNone` is useful - functions that return a bool - then the proposed functions are useful too. 

I’m also a big fan of falling into the pit of success. I don’t think that means minimalist and only one way to do something. It means finding a good balance. Technically we don’t need `mapOr`. We could just have `map` followed by `getOr` But it is convenient to combine them. I think the same applies with the functions here.

PR #131, opened 2023-04-14
docs(option): getUnsafe does not raise an exception

### LINKED ISSUE #130: Option.getUnsafe does not raise
The documentation of `Option.getUnsafe` gives the example:

> Option.getUnsafe(None) // Raises an error

However, the following code does not raise. I see `undefined` printed to the console and no exception.

```
let x = Option.getUnsafe(None)
Js.log(j`$x`)
```

Why is there also an `Option.getExn` function? That has the expected behavior, so I'm using that for now.

glennsl:
Fixes #130

### DIFF
--- a/src/Core__Option.resi
+++ b/src/Core__Option.resi
@@ -80,7 +80,7 @@ Option.getExn(None) /* Raises an Error */
 let getExn: option<'a> => 'a
 
 /**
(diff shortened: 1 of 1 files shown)


zth (approved):
Thanks @glennsl !

PR #136, opened 2023-04-27
make String.localeCompare return an int

tsnobip:
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#description) states that localeCompare "returns an integer indicating whether the referenceStr comes before, after or is equivalent to the compareString."

### DIFF
--- a/src/Core__String.res
+++ b/src/Core__String.res
@@ -130,4 +130,4 @@ external splitByRegExpAtMost: (string, Core__RegExp.t, ~limit: int) => array<opt
 @get_index external getSymbolUnsafe: (string, Core__Symbol.t) => 'a = ""
 @set_index external setSymbol: (string, Core__Symbol.t, 'a) => unit = ""
 
-@send external localeCompare: (string, string) => float = "localeCompare"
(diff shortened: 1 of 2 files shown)


glennsl:
The specification just says it's a number, and that the actual return values are implementation-defined: https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.localecompare

tsnobip:
ok closing then

jmagaram:
Whenever I need to sort by string I just want to be able to plug in String.compare and not have to write my own ternary. Most places in ReScript, like Array.sort and Belt set, expect an int returning function. Can we have a version of this that returns an int? Will something break if we just coerce the return value to an int?

glennsl:
I'm increasingly of the opinion that compare functions should all use `float`s. One example I just stumbled on is comparing dates, where the simplest and most performant method is to subtract their timestamps. But since those values are larger than rescript `int`'s, one cannot simple convert them, or the subtraction result to `int`s.

tsnobip:
I think it would totally make sense.

PR #139, opened 2023-05-22
Use idiomatic new JS array immutable helpers

### LINKED ISSUE #138: Sort out sort
This is coming: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted

chenglou:
Fixes #138

Added `toSorted`, `toReversed`, `toSpliced` and `with`, per spec.

Which means moving `sortInPlace` to `sort`. Same for `reverseInPlace` -> `reverse` and `spliceInPlace` -> `splice`.

This is a small breaking change, thankfully caught by the type system, and by Core typing these mutable helpers' return as `unit`.

Did the same for `TypedArray`. Though its `reverse` and `sort` _did_ return the array itself, so this breaking change should be carefuly advertized.

The `toBla-ed` naming convention seems to be the way to go in JS in the future. We should consider that for us as well. Otherwise we'd end up diverging more and more when JS releases more immutable helpers.

### DIFF
--- a/src/Core__Array.resi
+++ b/src/Core__Array.resi
@@ -165,7 +165,7 @@ Console.log(someArray) // ["hi", "hello", "yay", "wehoo"]
 external pushMany: (array<'a>, array<'a>) => unit = "push"
 
 /**
-`reverseInPlace(array)` reverses the order of the items in the array.
+`reverse(array)` reverses the order of the items in `array`.
 
 Beware this will *mutate* the array.
 
@@ -174,13 +174,13 @@ See [`Array.reverse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Re
 ## Examples
 ```rescript
 let someArray = ["hi", "hello"]
-someArray->Array.reverseInPlace
+someArray->Array.reverse
 
 Console.log(someArray) // ["hello", "h1"]
 ```
 */
 @send
-external reverseInPlace: array<'a> => unit = "reverse"
+external reverse: array<'a> => unit = "reverse"
 
 /**
 `shift(array)` removes the first item in the array, and returns it.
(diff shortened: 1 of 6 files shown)


src/Core__Array.resi
```diff
@@ -225,16 +227,21 @@ See [`Array.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Refer
...
+
+@variadic @send
+external splice: (array<'a>, ~start: int, ~remove: int, ~insert: array<'a>) => unit = "splice"
+
 @variadic @send
-external spliceInPlace: (array<'a>, ~start: int, ~remove: int, ~insert: array<'a>) => unit =
-  "splice"
+external toSpliced: (array<'a>, ~start: int, ~remove: int, ~insert: array<'a>) => unit = "toSpliced"
+
+@send external with: (array<'a>, int, 'a) => unit = "with"
```
glennsl, line 244:
```suggestion
@send external with: (array<'a>, int, 'a) => array<'a> = "with"
```

glennsl (changes requested):
Seeing as these functions haven't been implemented yet by all mainstream browsers (in particular Firefox), I would say this is quite a large breaking change for most consumers. There ought to be a guideline for what should be bound to, and whether it should be polyfilled.

Since this is existing functionality, I think it ought to be polyfilled for now to reduce breakage.

Also, while the change in type signature will warn of the breakage, because the difference is only in return type the actual errors might be located far away from its cause.

chenglou:
Ouf I think I committed a wrong draft. Thanks for catching these. Fixed.

I'll let others comment on these bigger philosophical decisions, but my personal opinions are:

> There ought to be a guideline for what should be bound to, and whether it should be polyfilled

This'd create more discussions & scheduling around when to un-polyfill, which would create another round of surprises. Plus Node in the future. We should specify that we're binding to version X of node for example, but for browsers I don't think we should be in the game of watching and communicating these. It's much less surprising if it's a uniform, consistent message that we'll bind to what's landing in browsers, but the user's codebase's still responsible for polyfilling, which they might already do.

> because the difference is only in return type the actual errors might be located far away from its cause

Yeah the error location is unfortunate, but I think that ship might have sailed when we decided to keep full inference instead of mandatory annotation at function boundary (to be clear I prefer the former); from here on it's all best-effort, and as long as it's communicated clearly in the changelog I'd say we'd done our job right.

(Btw there's the edge-case of polyfilling if we ever upgrade and change an API such that folks _now_ need to polyfill, in which case I think we should definitely polyfill. But that doesn't seem likely)

cknitt:
> The `toBla-ed` naming convention seems to be the way to go in JS in the future. We should consider that for us as well. Otherwise we'd end up diverging more and more when JS releases more immutable helpers.

Agreed. I am fine with these changes if they are clearly advertised in the CHANGELOG including a note regarding polyfilling.

glennsl (approved):

PR #141, opened 2023-05-22
Upgrade Test stack trace cleaning function

chenglou:
The stack format change slightly and we were no longer matching on the correct stuff to clean up: "(internal/..." became "(node:internal/..."

Took the occasion to swap them out with regex and inline the one-off cleanup logic too (IIRC I kept it separate because not all repos wanted that cleanup. But nobody cares now)

Before:
Screenshot 2023-05-22 at 03 43 30
After:
Screenshot 2023-05-22 at 03 43 08

### DIFF
--- a/test/Test.mjs
+++ b/test/Test.mjs
@@ -9,25 +9,6 @@ import * as CodeFrame from "@babel/code-frame";
 
 var dirname = (new URL('.', import.meta.url).pathname);
 
-function cleanUpStackTrace(stack) {
-  var removeInternalLines = function (lines, _i) {
-    while(true) {
-      var i = _i;
-      if (i >= lines.length) {
-        return lines;
-      }
-      if (lines[i].indexOf(" (internal/") >= 0) {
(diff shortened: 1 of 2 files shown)


chenglou:
Random note we should put somewhere more prominent:

So someone mentioned wanting to make this `Test.run(__POS_OF__("describe things"), ...)` a bit prettier, which is understandable. Though I wanna mention a nuance about the `__POS_OF__` part that folks are probably not leveraging. You can optionally carry that first-class, into a helper function, deep into its loop or whatever, and the stack trace will _still_ show your original test's location. For example, compare this:
Screenshot 2023-05-22 at 04 08 41
Screenshot 2023-05-22 at 04 08 24

To this, where `testHelper` receives `__POS_OF__` from its caller:
Screenshot 2023-05-22 at 04 09 08
Screenshot 2023-05-22 at 04 07 36

The second one's directly informative of where the call went wrong, which makes the code frame more useful. It also shows the proper test description

PR #143, opened 2023-05-28
Map, Set, WeakMap, WeakSet: use types from Js namespace

cknitt:
Placeholder types for ES6 collections were added to the Js namespace in https://github.com/rescript-lang/rescript-compiler/pull/5630.

We should use these in Core.


zth (approved):
Mind adding a changelog?

PR #148, opened 2023-06-30
Fix Math.atan2 argument order

### LINKED ISSUE #147: Argument order for Math.atan2 is reversed
From the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2):
```js
Math.atan2(y, x)
```

In the bindings:
https://github.com/rescript-association/rescript-core/blob/3a45914f499d543ceb21a1fe89618afd09692e3a/src/Core__Math.resi#L402-L403

Minnozz:
Fixes #147


zth:
Thanks!

PR #151, opened 2023-07-25
Clean up Result and Option modules

DZakh:


DZakh:
@zth No API changes, cleaning up some comments, removing unused code, improving compiled code

Issue #23, opened 2023-02-10
[Discussion] Immutable defaults for mutable JS API:s


zth:
Following the discussion about immutable defaults in https://forum.rescript-lang.org/t/ann-rescript-core-a-new-batteries-included-drop-in-standard-library/4149, let's discuss which API:s are problematic and what we can do to remedy that.

First off, there's a few things we need to keep in mind in this discussion:
- A goal is to stay close to the JS API:s (deliberately _not_ saying "as close as possible", but close). This means we have wiggle room to brush over inconsistencies and "improve" the JS API:s we bind to so they make more sense for idiomatic ReScript usage, where it makes sense. But a complete overhaul of API:s isn't something we'll pursue for `Core`. Not everyone is going to agree with this, and that's fine. But that's the way forward for `Core`.
- For full immutability, `Belt` is still a great alternative, and `Belt` isn't going anywhere, it'll keep being around.

With that said, providing immutable defaults where it makes sense is definitively a good idea. So, let's detail what API:s are problematic right now. Please write down the API:s with mutability you worry about, and we'll discuss what we can do about it, if anything.

Let's also discuss what we can do about the mutable API:s to make it apparent that they're mutable. A few things come to mind:
- Return `unit` always, even if the underlying JS API returns itself mutated. One example for this is `Map.set` which mutates and returns itself. This would return `unit` instead to signify that it's mutating.
- Naming. This is already the case, suffixing functions mutating the original thing with `inPlace`. We can check that that's consistent.
- Doc strings. Have a bolded `**This mutates the original array**` on top of each doc string, plus suggesting the immutable alternative. This will make it very apparent that you're using a mutable API as you're working in your editor.

DZakh:
Also, there's https://github.com/tc39/proposal-change-array-by-copy. The question is how far we should go from the JS API

hoichi:
From perspective of a ReScript module API, making dangerous functions more visible is good, but I’m not sure if it’s good enough. When a module `Foo` has no mutating methods, you have a strong guarantee that `Foo.t` is immutable, and the invariant _new value == new ref_ is enforced. Which have consequences for pattens that take advantage of this invariant, e.g. the React hooks with dependencies arrays.

[...]

PR #152, opened 2023-07-28
ImmutableArray

jderochervlk:
An addition to add an `ImmutableArray` type to Core.

Conversation over here: https://github.com/rescript-association/rescript-core/issues/23

This just omits any functions that would mutate the original Array and adds a function to create an `ImmutableArray` from an `Array`.

Here's an example of how to create one:
```rescript
let a1 = [1, 2, 3]

let immutableA1 = RescriptCore.ImmurableArray.fromArray(a1)
```

And here's the output:
```js
// Generated by ReScript, PLEASE EDIT WITH CARE


var a1 = [
  1,
  2,
  3
];

var immutableA1 = a1.slice();

export {
  a1 ,
  immutableA1 ,
}
/* immutableA1 Not a pure module */
```

I made `fromArray` use `slice` to create a copy of the original array. This means that once we create an `ImmutableArray` from a normal `Array` we don't have to worry about the original array mutating. Of course we only have that guarantee that no one will mutate the array if all of our source code is written in Rescript. 

I'll address these items if this gets the go ahead to be added.
TODO:
- [ ] update and write docs
- [ ] write tests


jmagaram:
JavaScript and .net combine both mutable and immutable array functions in the same module because it is convenient for the programmer and has performance benefits. The programmer can use immutable functions most of the time and carefully switch over to a mutable "push" for performance benefits. I think arrays are mutable almost by definition. I like immutability, but just don't think Array is a natural fit for it.

If we have such a module that represents data structures NOT in JavaScript I think other data structures are higher priority or more useful. For example the Belt Set and Map allow you to define the item equality operator, without which the JavaScript Map and Set are somewhat useless. Here are immutable collections for .net - lots of interesting ones to choose from. https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable?view=net-7.0 Personally I find a NonEmptyArray to be useful. And finally I think lazy sequences are extremely useful, especially for working with arrays, are immutable, and are part of most mature frameworks. I think sequences are far higher priority than immutable arrays. https://github.com/jmagaram/rescript-seq
So if we're going to have immutable arrays I think this should be part of a larger discussion about data structures. 

As for this particular code, once you've got an array it seems like you'd need a `toArray` so you could consume it with most of the code out there that requires one. I would expect this to be zero cost and introduce the risk that some other code will change it.

jderochervlk:
> As for this particular code, once you've got an array it seems like you'd need a `toArray` so you could consume it with most of the code out there that requires one.

Correct. This would mean you would have to convert this to a regular `Array` before you could pass it to `React.array` for example. This type of moving around can be a pain, but it's no different than using Immutable.js where that type of workflow is common.

Ideally this would be similar to the `ReadonlyArray` type from TypeScript which doesn't allow the functions that are removed in this PR. You also can't pass a `ReadonlyArray` to a function that expects an array, but in TS you can also take in `a: ReadonlyArray<int> | Array<int>`.

https://www.typescriptlang.org/play?#code/DYUwLgBGCMBcECUQEMAmB7AdsAngQQCcDkcAeTAVwFsAjEAgPggF4IBtaAGgCZOBmALoAoIQDMKmAMZgAllggypBEFRCYwACmTxCxMpVr0GASggBvIRCsRlYCgUwRkAOirIADhsfMmjgNQQ0MZCAL4ioJDKAM4UwJCsipLKquoaMMFiEtJyjqKgAB4yNKAAkkoqapraELok5NR0jBAAPogoGNj4RHUGjSbmlta29o4ubp7evhABQaHh4BBRIJJYqEgxcSwQeSCFxSBlSRWp6UA

cristianoc:
Notice casting one type of array to the other (either direction) without copying is unsafe.
Also, immutable array can be `+'a` so it works with sub typing.

cristianoc:
In case one wants to intentionally be unsafe (immutable arrays can actually be mutated by code that type checks), copying TS behavior, one would need to explain the benefits of the choice.

jderochervlk:
> Notice casting one type of array to the other (either direction) without copying is unsafe.
> Also, immutable array can be `+'a` so it works with sub typing.

Correct, in order to move from a mutable array from immutable, or the other way around, the array is copied.

> In case one wants to intentionally be unsafe (immutable arrays can actually be mutated by code that type checks), copying TS behavior, one would need to explain the benefits of the choice.

TypeScript does it's best to try and stop you from mutating a readonly array.

You can't do this for example:
```ts
function fn(a: Array<number> | ReadonlyArray<number>) {
    return a.push(1) // Property 'push' does not exist on type 'readonly number[]
}
```

TypeScript also doesn't let you swap the type from a `ReadonlyArray` to an `Array`, you need to make a copy.
```ts
function convert<T>(a: ReadonlyArray<T>): Array<T> {
    return [...a]
}
function convert<T>(a: ReadonlyArray<T>): Array<T> {
    return a //The type 'readonly T[]' is 'readonly' and cannot be assigned to the mutable type 'T[]'.
}
```

cristianoc:
https://www.typescriptlang.org/play?#code/FAMwrgdgxgLglgewgAikgbgUwE4wBTIAeyAXMgILbYCGAngDwDOM2cEA5gHwCUpyASpmoATJABtalGg2asOnZAG9gyVcmyYYYbCkLAAvsCA

jderochervlk:
Yeah, TypeScript isn't perfect. Hopefully this is something that can enforced going in both directions with ReScript.

PR #126, opened 2023-03-29
Array.findLast and other missing functions

jmagaram:
Some missing functions from the Array module. In particular the `findLast` variants.

`indexOfFromOpt`: add code and docs
`lastIndexOf` : docs, tests
`lastIndexOfOpt` : tests
`lastIndexOfFrom` : docs, tests
`findLast` : code, docs, tests
`findLastIndex` : code, docs, tests
`findLastIndexOpt` : code, docs, tests
`findLastIndexWithIndex` : code, docs, tests
`findLastIndexWithIndexOpt` : code, docs

PR #153, opened 2023-09-28
chore(ci): update node matrix

glennsl:
Removes 16.x, which is now unsupported.
Replaces 19.x with 20.x as the current release

This should unblock #126


zth:
👍

PR #154, opened 2023-10-10
Add support for more code conversion for migration

seekshiva:


zth:
This is great, thank you! Don't hesitate to add more if you find.

zth:
@seekshiva is it ready to be merged?

seekshiva:
@zth This was required to convert one of our internal repos into rescript. If we find more such cases while converting other repos, will raise more PRs in future.

That's it for now.

seekshiva:
I also found that `Js.Math.random_int()` no longer exists. We should either prevent this from getting converted into `Math.random_int()` or we need to add `random_int` into the Math module.

I wasn't sure which direction to go in. If you have suggestion on this, I can implement it in the same PR or in a different one.

PR #156, opened 2023-10-19
Added random_int to Core__Math

adhi-27:


DZakh:
Hi, it should be added in the module https://github.com/rescript-association/rescript-core/blob/106a882283806ec130e37a1a8781f0b0c41578e5/src/Core__Math.res#L12

zth:
Looking good! We'll need docstrings and tests for this too.

adhi-27:
> Looking good! We'll need docstrings and tests for this too.

Added docstrings and tests

zth (approved):
Great stuff, thank you! Mind adding a changelog too?

adhi-27:
> Great stuff, thank you! Mind adding a changelog too?

Added the changelog

zth:
@adhi-27 I think the changelog entries ended up in the wrong place. They should be directly under the "main" entry at the top. Looks like there are no current entries there, so you'll also need to add the sub headlines (Documentation etc).

PR #157, opened 2023-10-27
Only allow to join actual strings with Array.joinWith

fhammerschmidt:
The old `joinWith` binding sometimes led me to write code that resulted in `[object Object]` which cannot happen with this stricter signature. Not sure if the unsafe variant is really needed.


zth:
Great! Changelog and then we're good to go.

PR #162, opened 2023-11-18
Rescript v11

cwstra:
I was playing around with Rescript 11 + core, and noticed the breaking JSON type change wasn't added. Seemed like a simple enough thing to tweak myself, so here we are.

There are two parts of the tests that seem to cause issues, though presumably they have nothing to do with my changes:

- `PromiseTest.res`, line 226, throws an exception, as it tries to use an undefined variable `v$1`.
- `TestTests.res`, line 20-25, fails normally.

Both of these seem to be due to an underlying compiler change; namely, if the sole argument to a function is `unit`, the function arguments of the resulting JS will be empty. I'd imagine this somehow spills out of uncurried mode, since unit arguments are more common in curried arguments.


zth:
Thank you @cwstra ! I'll get back to this one soon. After v11 is released we'll focus Core on v11 entirely, which will open up some interesting changes.

zth:
Superseded by the dedicated v11 PR.

PR #163, opened 2023-11-27
Add String.replaceAll and String.replaceAllRegExp

fhammerschmidt:
While replaceAllRegExp is not that useful since you can just do the normal one with the `/g` flag, the string one is a very welcome addition imo.


zth:
Excellent, have been thinking about these as well! Ready to go?

fhammerschmidt:
Yes.
I just wished we had a way to check at compile time that a `g` flag is provided in `replaceAllRegExp`. Runtime errors are not really the ReScript way.

Issue #170, opened 2023-12-14
`Nullable.fromOption` does not return `null` for `None`


jquesada2016:
```rescript
None->Nullable.fromOption
```

should return `null`, but instead returns `undefined`, which is not the same. :(

zth:
This is actually intended. Maybe it's the naming that's a bit confusing, but `Null.t` and `Null.fromOption` would return `null` for `None`. `Nullable` has both `undefined` and `null`.

Closing since it's working as intended, but please feel free to continue the discussion.

DZakh:
Related to https://github.com/rescript-lang/rescript-compiler/issues/6533

jquesada2016:
Yes, this was exactly what got me. The `nullable` name made me assume it returned `null`, rather than it being a nomenclature fore both `null` and `undefined` coercing to `undefined`. I also thought `null` only defined the `null` constant value. I hadn't noticed `Null.t<'a>`.

PR #171, opened 2023-12-14
removed `undefined` from `Core__Nullable`

jquesada2016:
This PR addresses #170. It simply removes `external undefined` from the `Nullable` module.

### DIFF
--- a/src/Core__Nullable.mjs
+++ b/src/Core__Nullable.mjs
@@ -7,8 +7,9 @@ import * as Core__Option from "./Core__Option.mjs";
 function fromOption(option) {
   if (option !== undefined) {
     return Caml_option.valFromOption(option);
+  } else {
+    return null;
   }
-  
 }
 
 function equal(a, b, eq) {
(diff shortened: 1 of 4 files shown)


DZakh:
Undefined is part of nullable. There's Null.t if you dont't need undefined

jquesada2016:
Wouldn't this be more clear if we had an `Undefinable` module to seperate `null` from `undefined`. I've run across this issue working with `JSON.stringifyAny` where one of the types is `Nullable.t<'a>`, and it skipping the field rather than putting `null` when the value is supposedly null.

illusionalsagacity:
> Wouldn't this be more clear if we had an `Undefinable` module to seperate `null` from `undefined`.

If you want strictly undefined or `'a` you probably want an `option<'a>` instead as that sounds like it'd match your expectations here; and I don't think introducing a breaking change to Core when this matches the behavior of the old Js module would help with the migration path

> I've run across this issue working with `JSON.stringifyAny` where one of the types is `Nullable.t<'a>`, and it skipping the field rather than putting `null` when the value is supposedly null.

That sounds like it matches the JavaScript implementation to me:

```javascript
JSON.stringify({hello: undefined, world: "foo"}) // outputs '{"world":"foo"}'
```

PR #184, opened 2024-01-22
Add Tests for Doc Examples

aspeddro:
- Add a small test script to test docstrings examples
- Docstring examples must be compiled
- Test with ReScript v11


zth:
Very nice! A few notes/thoughts:

- I tried it and it's quite slow. I guess a fast win could be to make it run in parallell? Chunked by CPU cores minus one perhaps.
- Can we name the temp file after the file it was found in instead + what value/type it is attached to?
- I wonder if using `@rescript/tools` and the docs extraction is a better base for this than scanning the folders manually. What do you think?
- Let's think about how we can generalize this in the future too. Maybe this functionality could ship together with `@rescript/tools` so it's easy to set this up in your own generated docs.
- Related to the above, a cool future development would be to have it autoinject the generated JS into the docs, if wanted.

An alternative approach that perhaps could solve the above issues could be to:
- Emit and instantiate a test project like you do now (could reference the project locally in `package.json`, doesn't need to be packaged up I think)
- Use `@rescript/tools` docgen to find all example code. Emit into files in the example project, like: `Core__Array__reduce.res` for `Core__Array.reduce`
- Build the project. This should let ReScript build it all in parallell. All the info we want to show should be in stderr, and we could simply check if the compiler exits successfully or not to see if CI should be stopped.

What do you think?

aspeddro:
Ready for review.

The commit https://github.com/rescript-association/rescript-core/pull/184/commits/3c4f07dd5cd8168ed35c0ed779c93b6484a541fe has the script test

https://github.com/rescript-association/rescript-core/pull/184/commits/7fe04fbd8cd0f3772b9acb72f0a900cb2ac56be7 fix for all docstrings reported [here](https://github.com/rescript-association/rescript-core/actions/runs/7952166500/job/21706426501#step:6:1)

aspeddro:
Ok. This is ready for review.

Two commits here. Fist https://github.com/rescript-association/rescript-core/pull/184/commits/f3af423b7e127186a39144c90cebaf989ad23d86 commit has the script tests using `rescript/tools` to extract docstrings

The new error log:
image

See all errors reported [here](https://github.com/rescript-association/rescript-core/actions/runs/7983348768/job/21798360285)

The last commit https://github.com/rescript-association/rescript-core/pull/184/commits/89b399bbf7fd5c0ab2d2fd3b2e647800098cf599 fixes all docstrings examples.

zth (approved):
This is great! We can refactor the script later on if we want, for now this works well.

PR #188, opened 2024-01-27
add ci workflow: JS files are up to date

### LINKED ISSUE #182: CI: check whether the committed compiled JS files are up-to-date
It is possible to submit a PR that changes the `.res`/`.resi` files but not the corresponding `.mjs` files, for instance if you forget to run the compiler after your last change. I think this should be checked in CI.

- Run the compiler
- Check that there are no changes compared to the committed files

aspeddro:
Close #182

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -1,0 +1,22 @@
+name: Check JS Files
+
+on:
+  push:
+    branches: [main]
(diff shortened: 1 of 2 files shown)


zth:
Perfect! Will you update the offending file so it passes too?

aspeddro:
Done

PR #191, opened 2024-02-15
Get rid of U functions

cknitt:
Not needed anymore in uncurried mode.


zth:
@cknitt worth a changelog entry perhaps?

cknitt:
> @cknitt worth a changelog entry perhaps?

done

PR #193, opened 2024-02-15
Rename Object.empty to Object.make

glennsl:
`Dict` and every other module uses `make` to create empty collections. For consistency, `Object` should do the same.


zth:
Agreed. Thanks!

PR #195, opened 2024-02-17
Align List api with other modules

cknitt:
Align `List` api with `Array` and other modules:

* `List.make`: named argument `length`
* `List.makeBy` -> `List.fromInitializer`, named argument `length`
* `List.flatten` -> `List.flat`
* `List.mapWithIndex`: callback argument order
* `List.forEachWithIndex`: callback argument order, callback returns `unit`
* `List.getBy` -> `List.find`


zth:
@cknitt changelog, and what do you think about adding this to the migrations script as well, if that makes sense?

PR #197, opened 2024-02-24
fix(intl): expose Intl.Common types

glennsl:
The Intl.Common types don't seem to be exposed. And while they technically don't need to be, since they're structural, it's useful to be able to reference them for derived APIs (as I have in a project of mine).

It might be more consistent to expose the types directly in relevant modules though, instead of the module as a whole, but this is the minimal change needed.


zth:
@glennsl perfect, thank you for cleaning up as well! Would you like a `1.1.1` with this so you can use this in your project right away? Easily done if so.

glennsl:
Thanks, but no need for my sake. I found the `zeroTo20` and `zeroTo21` types a bit too restrictive amyway, so I just continued to use `int`s and sprinkled some magic on them instead.

PR #199, opened 2024-03-01
Add Array.flatMapWithIndex

mediremi:
[`Array.prototype.flatMap` passes the current element's index to its callback argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#callbackfn) but we do not have a `Array.flatMapWithIndex` function.


zth:
Seems reasonable, thanks! Would you add a changelog as well?

mediremi:
@zth I've added a changelog entry :+1:

PR #201, opened 2024-03-08
Use optional args for JSON.stringify and JSON.parseExn

cknitt:
This would simplify the API a lot, but it would be a breaking change.

What do you think?

### DIFF
--- a/src/Core__JSON.resi
+++ b/src/Core__JSON.resi
@@ -36,15 +36,12 @@ try {
 | Exn.Error(_) => Console.log("error")
 }
 
-let reviver = (_, value) => {
-  let valueType = JSON.Classify.classify(value)
-
-  switch valueType {
+let reviver = (_, value: JSON.t) =>
+  switch value {
   | String(string) => string->String.toUpperCase->JSON.Encode.string
   | Number(number) => (number *. 2.0)->JSON.Encode.float
   | _ => value
   }
-}
 
 let jsonString = `{"hello":"world","someNumber":21}`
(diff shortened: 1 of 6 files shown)


zth:
Love it! If we keep the old functions around for a while but deprecate them then we can avoid breaking, right?

cknitt:
> Love it! If we keep the old functions around for a while but deprecate them then we can avoid breaking, right?

Yes, will do that! (And also update the documentation and add some tests.)

cknitt:
@zth BTW, [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#space) and TypeScript typedefs call the last parameter `space` instead of `indent`. I think we should do that, too.

cknitt:
@zth Ready for re-review, could you have a look?

src/Core__JSON.resi
```diff
@@ -31,6 +35,28 @@ try {
 } catch {
 | Exn.Error(_) => Console.log("error")
 }
+
+let reviver = (_, value) => {
+  let valueType = JSON.Classify.classify(value)
```
zth, line 40: Does this really need classify? Can't it use the Json representation directly?

zth (approved):
This looks good to me. 2 comments:
- Docstrings should probably be updated to match the new v11 Json representation.
- Will this require 11.1 to play nicely with undefineds?

zth (approved):

PR #202, opened 2024-03-16
Add Option.all & Result.all helpers

bloodyowl:
Adds `Option.all` & `Result.all` helpers. The API is inspired by `Promise.all` which is pretty well known by users.

```res
Option.all([]) // Some([])
Option.all([Some(1), Some(2), Some(3)]) // Some([1, 2, 3])
Option.all([Some(1), None]) // None

Result.all([]) // Ok([])
Result.all([Ok(1), Ok(2), Ok(3)]) // Ok([1, 2, 3])
Result.all([Ok(1), Error(1)]) // Error(1)
```


cknitt:
I think that if we want to add this, we should also have `all2`, `all3`, ... like we do for `Promise`.

(This is something that I would actually already have needed in practice for the `Result` module.)

test/intl/Intl__CollatorTest.mjs
```diff
@@ -24,7 +24,7 @@ var collator = new Intl.Collator(["en-US"], {
 Intl.Collator.supportedLocalesOf([
       "en-US",
       "en-GB"
-    ], undefined);
+    ]);
```
cknitt, line 27: There are some unrelated changes in the test output like here - were you using a different/newer ReScript version?

bloodyowl:
closing as merged with https://github.com/rescript-lang/rescript/pull/7181

PR #203, opened 2024-03-17
Remove incorrect asr binding and add lsr

cometkim:
JS BigInt doesn't support ASR (`>>>`)


cometkim:
It's opposite. I was misunderstanding 😥

PR #205, opened 2024-04-01
`Array.joinWith` -> `Array.join`

cknitt:


zth:
I think this might be a bit problematic. While more correct, we're making the most common use case by far, which is passing an arg, more cumbersome to do.

I think we should just rename to join and have the arg still be an unlabeled string. That's the most common operation.

cknitt:
Fine with me, too! Will update the PR.

cknitt:
@zth Done! Please re-review. And could you also have a look at #201?

zth (approved):

PR #207, opened 2024-04-02
Use native bigint type

fhammerschmidt:


zth:
This is great! But let's do a normal release first so we can bump to 11.1 for the next one. I'll sort it out soon.

fhammerschmidt:
actually adapted it to be basically the same as Js.Bigint, since now we can do (and document) just a 
`open! BigInt` when Core is open.

fhammerschmidt:
With rc-7 the doc tests succeed!

zth:
I'll make a new release soon and we can get this merged after.

zth:
@fhammerschmidt rebase and we're ready to go!

PR #210, opened 2024-04-07
PoC: Runtime tests using examples from docstrings

aspeddro:
This PR is a proposal to add unit tests using docstrings. Currently the examples from docstrings have tests for compilation only.

- `bsc` emits the Js and passes the code to `node` using the `--eval|-e` argument.

We need some functions to deal with errors in a idiomatic way

Something like this
```rescript
List.tailExn(list{1, 2, 3})->assert_eq(list{2, 3})

List.tailExn(list{})->Exn.isExn
```

image


aspeddro:
Ready for review

fhammerschmidt (approved):
really nice work, some of the examples could be a bit more beginner-friendly but I know that is very hard to do without making them untestable.

aspeddro:
I made a small improvement in two examples in the `Null` and `Nullable` module

fhammerschmidt (approved):

fhammerschmidt:
@zth what do you think?

PR #216, opened 2024-04-21
Fix signature for Array.splice

bloodyowl:


leoliu:
Consistency with JS would be nice.

fhammerschmidt:
Ah I think we tried to return unit for mutating methods, but splice is a bit special since it returns the other half of the array.
@zth WDYT?

PR #217, opened 2024-05-03
:bug: - Fix binding of Array.get, to make sure it boxes options

jfrolich:
This can cause runtime crashes when the individual members are options. When you are using `Option.getExn` after an `Array.get` to get the value if you know the array is in bounds, it will crash if the item in the array an option type and is `None` for the element. In normal cases you would use `Array.getUnsafe` for this, but in more complex scenarios this can happen. The types are actually incorrect and it can lead to crashes where you wouldn't expect it.

### DIFF
--- a/src/Core__Array.mjs
+++ b/src/Core__Array.mjs
@@ -99,6 +99,13 @@ function reduceRightWithIndex(arr, init, f) {
   return arr.reduceRight(f, init);
 }
 
+function get(arr, i) {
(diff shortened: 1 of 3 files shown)


glennsl (commented):
This would also apply to every other function that returns `option<'a>`, wouldn't it? Such as `at`, `pop`, `shift`, `find` etc.?

zth:
@jfrolich ping about @glennsl comment

PR #218, opened 2024-05-03
Add `Math.Int.ceil`

jfrolich:
Also fixes the documentation of `Math.Int.floor` since it's incorrect


zth:
@jfrolich could we get a changelog?

Issue #220, opened 2024-05-15
@rescript/core does not support ESM.


htoooth:
When I compile rescript into JavaScript and have Node.js run it, I found that the `@rescript/core`was not set as a esmodule. When I manually changed the `package.json` in the `core` to `type: module`, everything worked properly.

```bash
(node:46659) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/projects/rescript-message/node_modules/.pnpm/@rescript+core@1.3.0_rescript@11.1.0/node_modules/@rescript/core/src/Core__List.res.js:3
import * as Belt_Array from "rescript/lib/es6/belt_Array.js";
```

jderochervlk:
You need to set the file extension to `.mjs` or `.res.mjs` to opt into ESM for dependent packages. If we were to set `"type": "module"` in the package.json for Core it would mean ESM would be the default and for CJS support you would need to set the file extension to `.cjs`. Opting into ESM is probably the best path now, but the docs could probably be clearer around this. Soon this library will be part of the compiler, so it will be easier to keep everything aligned with ESM/CJS.

htoooth:
@jderochervlk  Thank you. You've saved me time.

tsnobip:
now that core is part of the compiler I guess we can close this one.

PR #222, opened 2024-05-18
Add section to README explaining settings to support ESM and CJS

jderochervlk:
Related to https://github.com/rescript-association/rescript-core/issues/220


zth:
Awesome, thank you!

PR #224, opened 2024-05-25
Correct comparison operator in Array.every() method

WooWan:
Fix typo in Array.every() comparison operator
Yay, my first contribution
```rescript
let array = [1, 2, 3, 4]

//before
Console.log(array->Array.every(num => num > 4)) // true

//after
Console.log(array->Array.every(num => num <= 4)) // true
```


cknitt (approved):
Thanks a lot! 👍

PR #229, opened 2024-06-05
add `Char` module

### LINKED ISSUE #228: Add `Char` module

aspeddro:
Perhaps add `chrOpt`, return `option<chr>`, @zth, @fhammerschmidt?

TODO:
- [ ] Get rid of `PervasivesU`
- [ ] Add docstrings
- [ ] Update CHANGELOG.md

Close #228


src/Core__Char.resi
```diff
@@ -0,0 +1,57 @@
...
+    {!Set.Make} and {!Map.Make}. */
+let compare: (t, t) => int
+
+/** The equal function for chars.
+    @since 4.03.0 */
+let equal: (t, t) => bool
+
+/* The following is for system use only. Do not call directly. */
+
+external unsafeChr: int => char = "%identity"
```
aspeddro, line 57: Rename to `chrUnsafe`

fhammerschmidt:
Yes, a `chrOpt` that checks if a given int is in the valid char range and only then returns a `Some(char)` is missing currently, imo.

PR #230, opened 2024-06-08
Remove `Exn` module

aspeddro:
The main reason is that all functions in `Js.Exn` are in `Error` module.


aspeddro:
Ready for review

PR #231, opened 2024-06-08
Remove `Re` module

aspeddro:
The `Re` and `RegExp` modules have the same target `Core__RegExp`


aspeddro:
Actually, I think it was intentional to maintain compatibility with `Js.Re` module, right?

cknitt:
Removed as part of #237.

Issue #232, opened 2024-06-11
RegExp.Result.t is unsafely typed for optional capture groups


illusionalsagacity:
see https://forum.rescript-lang.org/t/regexp-result-t-inconsistency/5326

https://github.com/rescript-association/rescript-core/blob/0b4bd00f518430f41bdfc8703b48191b458b4ded/src/Core__RegExp.res#L4

`Js.String2.match_` [here](https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/others/js_string2.res#L442) is typed as returning `option<array<option<t>>>`

PR #233, opened 2024-06-11
fix: RegExp.Result.t should be an array of options

illusionalsagacity:
Backport for #232


zth:
Nice! Could you add a changelog too?

PR #234, opened 2024-06-11
fix: RegExp.Result.t should be an array of options

### LINKED ISSUE #232: RegExp.Result.t is unsafely typed for optional capture groups

illusionalsagacity:
resolves #232

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next version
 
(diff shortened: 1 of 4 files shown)


zth (approved):
Nice!

PR #244, opened 2024-08-27
Add Bool module

fhammerschmidt:
As those methods will be removed from pervasives.
Also added some helpers inspired from https://hexdocs.pm/gleam_stdlib/gleam/bool.html

Please tell me which ones you would like to keep, then I will add tests and docs.


bloodyowl:
I'm wondering if we really need such a module in the stdlib, I'd rather encourage using plain expressions (inside anonymous functions if needed)

fhammerschmidt:
Probably not. It was mainly to have a substitute for the corresponding Pervasives methods, but we also have this now: https://github.com/rescript-lang/rescript/blob/master/runtime/Primitive_bool.res

PR #245, opened 2024-09-05
Int: Add bitwise functions from Pervasives

fhammerschmidt:


fhammerschmidt:
@cometkim fyi

cometkim:
BigInt needs this too

fhammerschmidt:
BigInt has a lot of missing docs so I will do that in a separate PR.

We also have the chance here to rename the functions into more descriptive ones, if we want.

lsl => logicalShiftLeft
lsl => logicalShiftRght
lxor => logicalXor
etc.

if we later on have operators for those functions anyway it would not matter as much if the names are longer.

cometkim:
> We also have the chance here to rename the functions into more descriptive ones, if we want.

We can "add", not rename

Even we support infix ops, the shorter names are still useful for those who prefer the pipe style.

fhammerschmidt:
@zth As mentioned in discord, I just keep it as it is. Merge?

zth:
Got for it!

zth:
Got for it!

fhammerschmidt:
wait, I need to put them in their own submodule first

PR #247, opened 2024-10-15
Add Set functions

andriijas:
Adds these missing functions

 Set.difference
 Set.intersection
 Set.union
 Set.symmetricDifference
 Set.isSubsetOf
 Set.isSupersetOf
 Set.isDisjointFrom
 
 Also adds convinience function `Set.toArray`  which lets you get the array of the set directly
 
 Before (still valid):
```
arr->Set.fromArray->Set.values->Array.fromIterator
```

New, shorter:

```
arr->Set.fromArray->Set.toArray
```

Creating a set from an array and then get the array of the set is a common pattern in javascript to remove duplicates from an array

ie 

```
  let set = Set.fromArray(["apple", "orange", "apple", "banana"])
  set->Set.toArray // ["apple", "orange", "banana"]
 ```
 
 
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next version
 
+- Add `difference`, `intersection`, `union`, `symmetricDifference`, `isSubsetOf`, `isSupersetOf`, `isDisjointFrom`, `toArray` functions to `Set`. https://github.com/rescript-association/rescript-core/pull/247
+
 ## 1.6.0
 
 - Optimize compare and equal functions. https://github.com/rescript-association/rescript-core/pull/238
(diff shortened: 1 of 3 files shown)


zth:
😍 

@andriijas could you add a link to MDN in the docstring of each of the newer `Set` methods you're binding to here? Then ready to go!

andriijas:
@zth updated PR ✅

Note: purposely left out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/entries and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/keys

I dont think those make sense on `Set`, I guess some javascript guy at tc39 thought it would be nice convenience to have same methods as on `Map`

zth (approved):
Fantastic!

PR #249, opened 2024-12-02
fix: incorrect comment on Math.sign

JUSTIVE:
image

Math.sign(negative_value) should be -1.0 instead of 1.0


cknitt (approved):
Thanks a lot!

rescript-core is now already integrated in the compiler repo for ReScript 12.
Could you do the same fix there?
https://github.com/rescript-lang/rescript/blob/master/runtime/Math.resi

PR #251, opened 2025-02-20
docs: updates the link in Math.ceil's docstring to point at the right page

### LINKED ISSUE #250: Doc string for Math.ceil links to Math.floor
https://github.com/rescript-lang/rescript-core/blob/f042582a9067a524e106250bdd20f970bafc21cc/src/Core__Math.resi#L295-L307

illusionalsagacity:
resolves #250


zth (approved):
Thanks!

PR #259, opened 2026-04-22
docs: fix doc comment bugs in List, Promise, and Float

mikaello:
Fix several documentation bugs found in doc comments:

- `List.fromInitializer`: description referenced old name `makeBy` (same as https://github.com/rescript-lang/rescript-core/pull/255)
- `List.reduceReverse2`: description said `reduce2` returns the final value
- `Promise.all6` / `allSettled6`: parameter list was missing `p3` (had `p1, p2, p4, p5, p6`), and the comment closing was malformed (`")*/`)
- `Float.Constants.maxValue`: example showed `minValue` instead of `maxValue`
- List (forEachWithIndex): Item 2 is cc → Item 2 is c (copy-paste typo)
- RegExp (sticky + unicode): example called `RegExp.unicode` instead of `RegExp.sticky`, and unicode description said flag was y instead of u


fhammerschmidt:
Thanks! ReScript Core has moved into the [main ReScript repo](https://github.com/rescript-lang/rescript) again. So please apply your fixes there first.

It's just called Stdlib internally in order to make migrations easier (no name clashes).

